PR #102, opened 2016-09-28
Allow the user to customize where the input history is saved

### LINKED ISSUE #101: mrepl-history file
It would be nice if the location of this file was customizable. Additionally, constructing the path should probably use a string which is portable. Looking at the source, comint-input-ring-file-name is hard-coded with a non-portable string.

PuercoPop:
Not sure is using `expand-file-name` is appropriate but if Emacs expands the path appropriately we don't have to rely on how CL implementations handle ~/ in the path.

* sly-mrepl.el (sly-mrepl-history-file-name): Add new variable
(sly-mrepl-mode-map): Use sly-mrepl-history-file-name instead of
  hard-coded path

Closes #101

### DIFF
--- a/contrib/sly-mrepl.el
+++ b/contrib/sly-mrepl.el
@@ -81,6 +81,10 @@ emptied. See also `sly-mrepl-hook'")
 This variables behaves like `comint-preoutput-filter-functions',
 for output printed to the REPL (not for evaluation results)")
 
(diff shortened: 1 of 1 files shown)


marsam:
I think the current patch is incomplete those buffer local variables are set only on `sly-mrepl-mode` call, but `sly-mrepl--read-input-ring` responsible of reading the history is called on different places. Curiously I have a personal patch for this the https://github.com/marsam/sly/commit/99aa71d.

PR #143, opened 2018-01-27
delete window after selecting new window in pop-to-source-location

deepestthought42:
This fixes frequent 'Selecting deleted buffer' messages for me under
emacs 25.2.3.

Full disclosure: I have spent literally two minutes on this and have never looked into the sly/slime code in any  meaningful way. I am also not experienced in elisp. It does fix the error message and also makes kind of sense to keep as long as you depend on it.

### DIFF
--- a/sly.el
+++ b/sly.el
@@ -3538,14 +3538,14 @@ If called from an xref buffer, method will be `sly-xref' and
 thus also honour `sly-xref--popup-method'."
   (let* ((xref-window (selected-window))
          (xref-buffer (window-buffer xref-window)))
-    (when (eq method 'sly-xref)
-      (quit-window nil xref-window))
     (with-current-buffer xref-buffer
       ;; now pop to target
       ;;
       (select-window
        (sly--display-source-location source-location nil method)))
-    (set-buffer (window-buffer (selected-window)))))
+    (set-buffer (window-buffer (selected-window)))
+    (when (eq method 'sly-xref)
+      (quit-window nil xref-window))))
 
 (defun sly-location-offset (location)
   "Return the position, as character number, of LOCATION."


joaotavora:
Looks good, but I have to take a closer look at the implications.

It would help, though, if you could isolate the exact case which causes the
error message.

joaotavora:
There's something definitely not OK, since some tests are failing after your changes:

```
3 unexpected results:
   FAILED  find-definition-other-window-1
   FAILED  find-definition-other-window-2
   FAILED  find-definition-other-window-3
```

You can probably reproduce this with `make check-core` from the sly directory, supposing you have emacs 25.3 linked to the EMACS environment variable.

joaotavora:
I've tried some more and I can't reproduce it. It would be very useful if you would try to toggle debugging ON by using `M-x toggle-debug-on-error` and then working for a while with that. When the error pops up, send me the resulting backtrace and tell me briefly what you were trying to do.

deepestthought42:
Hey,

thanks for getting at this so quickly!  To reproduce it (without the
change I made) I simply do a sly-edit-definition on anything followed by
sly-xref-goto (by hitting enter) on one of the results. The backtrace
is:

Debugger entered--Lisp error: (error "Selecting deleted buffer")
  set-buffer(#<killed buffer>)
  (save-current-buffer (set-buffer xref-buffer) (select-window (sly--display-source-location source-location nil method)))
  (let* ((xref-window (selected-window)) (xref-buffer (window-buffer xref-window))) (if (eq method (quote sly-xref)) (progn (quit-window nil xref-window))) (save-current-buffer (set-buffer xref-buffer) (select-window (sly--display-source-location source-location nil method))) (set-buffer (window-buffer (selected-window))))
  sly--pop-to-source-location((:location (:file "/home/renee/quicklisp/dists/quicklisp/software/fare-quasiquote-20170124-git/packages.lisp") (:position 40) (:snippet "(defpackage #:fare-quasiquote\n  (:use #:uiop #:fare-utils #:common-lisp)\n  (:shadow #:list #:list* #:cons #:append #:nconc #:quote)\n  (:shadow #:kwote #:quotep #:n-vector #:make-vector)\n  (:documentation\n   \"Quasiquote implementation with and for pattern-m")) sly-xref)
  #[257 "\300\301\"\207" [sly--pop-to-source-location sly-xref] 4 "\n\n(fn LOCATION)"]((:location (:file "/home/renee/quicklisp/dists/quicklisp/software/fare-quasiquote-20170124-git/packages.lisp") (:position 40) (:snippet "(defpackage #:fare-quasiquote\n  (:use #:uiop #:fare-utils #:common-lisp)\n  (:shadow #:list #:list* #:cons #:append #:nconc #:quote)\n  (:shadow #:kwote #:quotep #:n-vector #:make-vector)\n  (:documentation\n   \"Quasiquote implementation with and for pattern-m")))
  apply(#[257 "\300\301\"\207" [sly--pop-to-source-location sly-xref] 4 "\n\n(fn LOCATION)"] (:location (:file "/home/renee/quicklisp/dists/quicklisp/software/fare-quasiquote-20170124-git/packages.lisp") (:position 40) (:snippet "(defpackage #:fare-quasiquote\n  (:use #:uiop #:fare-utils #:common-lisp)\n  (:shadow #:list #:list* #:cons #:append #:nconc #:quote)\n  (:shadow #:kwote #:quotep #:n-vector #:make-vector)\n  (:documentation\n   \"Quasiquote implementation with and for pattern-m")))
  sly-button-goto-source(#<marker (moves after insertion) in no buffer>)
  button-activate(#<marker (moves after insertion) in no buffer> t)
  push-button(140 t)
  push-button(return)
  funcall-interactively(push-button return)
  call-interactively(push-button nil nil)
  command-execute(push-button)

I can reproduce the test failure and indeed
sly-edit-definition-other-window (which seems to be what the test case
is about) opens another frame when it should open a window with the
change. Without the change is giving me the same error though.

Thanks, Renee.



João Távora writes:

joaotavora:
Hi @deepestthought42 , thanks for the backtrace. It is very interesting indeed, since your system acts as if the expression `(quit-window nil xref-window)` was rather `(quit-window t xref-window)`. You see, the second one kills the buffer where the choices are presented, while the first does not. But the first one is the one that is in the code!

I would ask you to do these simple things without your change:
1. After the crash, can you find a buffer named "*sly-xref for <lisp-name>*" anywhere in your buffer list?
2. Can you reproduce this behaviour by starting emacs with the `-Q` flag and loading just sly and nothing else?
3. In the case that the experiment  in 2 succeeds can you post your emacs configuration somewhere? There's something fishy there that's deleting a buffer when it shouldn't be.  Only the xref **window** should be deleted right after you press "enter" on your choice. I assume this to be in some package or customization that you are using in yoru configuration.

joaotavora:
Another couple of points

1. you wouldn't happen to have something like [this gem](https://superuser.com/a/466109) in your configuration would you?

2. Can you try with this change instead of yours?
```diff
diff --git a/sly.el b/sly.el
index 1fd748c3..5f52863f 100644
--- a/sly.el
+++ b/sly.el
@@ -3548,7 +3548,7 @@ thus also honour `sly-xref--popup-method'."
   (let* ((xref-window (selected-window))
          (xref-buffer (window-buffer xref-window)))
     (when (eq method 'sly-xref)
-      (quit-window nil xref-window))
+      (quit-restore-window xref-window 'bury))
     (with-current-buffer xref-buffer
       ;; now pop to target
       ;;

```

deepestthought42:
Hi João!

Thanks a lot for your effort! Yes, I have that gem in my configuration
and yes, your change gets rid of the problem.

Running 'make check-core'

now only shows:
Test find-definition\.3-1 passed unexpectedly

but it does that for me with all the variants I have tested.

Thanks!

p.s.: Anything particularly bad about the quit-window defadvice ?

João Távora writes:

joaotavora:
> Thanks a lot for your effort! Yes, I have that gem in my configuration
and yes, your change gets rid of the problem.

Hehehe, I knew it!

Well, I pushed a fix that will let you keep that bit of configuration and resolve your problem (though I think you should remove it).

> Test find-definition\.3-1 passed unexpectedly
> but it does that for me with all the variants I have tested.

That's normal. The CI system uses an older version of SBCL which fails the test, but yours and mine already pass it.

> Thanks!
Thank you for bringing this to my attention.

> p.s.: Anything particularly bad about the quit-window defadvice ?

Errr, yes :-). By advising functions you're effectively making them violate their contracts. In this case, quit-window is not supposed to kill the buffer with the default arguments, just bury it. But by advising it, you've changed that and broken all programs that depend on that behaviour (including SLY and, I quite possibly parts of Emacs itself).

I would recommend that you delete that advice and study alternatives:

1. Delete the advice and describe exactly why the default quit-window behaviour annoys you, and then use M-x report-emacs-bug to describe that, or ask in emacs-help@gnu.org about alternatives.
2. At least change the advice to this, though it still smells a little bit:
```lisp
(defadvice quit-window (before quit-window-always-kill)
  "When running `quit-window', always kill the buffer."
  (when (called-interactively-p 'interactive) (ad-set-arg 0 t)))
(ad-activate 'quit-window)
```

deepestthought42:
Yeah,

I just looked at what the advice actually does -- I had just blindly
copy&pasted it I think because of a lot of dired buffers -- and that is
indeed somewhat disturbing. I think the the solution is to rebind
e.g. 'q' in dired-mode to appropriate command ...

Sorry about that, Renee.

João Távora writes:

PR #189, opened 2018-10-01
Open the MREPL in another window when not visible

PuercoPop:
* contrib/sly-mrepl.el (sly-mrepl): Use switch-to-buffer-other-window
    instead of switch-to-buffer

This is related to #158. I find the behaviour of displaying the REPL in the current window undesirable. Normally I want to "open"/see a REPL in addition to the buffer I'm currently editing.

### DIFF
--- a/contrib/sly-mrepl.el
+++ b/contrib/sly-mrepl.el
@@ -895,11 +895,14 @@ arglist for the most recently enclosed macro or function."
   "Find or create the first useful REPL for the default connection.
 If supplied, DISPLAY-ACTION is called on the
 buffer. Interactively, DISPLAY-ACTION defaults to using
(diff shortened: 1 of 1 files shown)


PuercoPop:
Hey @joaotavora, reading the documentation of display-buffer-alist it appears I can use it to achieve the behaviour I want. I wasn't aware of this variable. Thanks for pointing me in the right direction :pray:

PuercoPop:
If we use display-buffer instead of switch-to-buffer then the following snippet does what I want. It appears switch-to-buffer is not affected by `display-buffer-alist`


```elisp
(setf display-buffer-alist
      (acons 
       (lambda (buffer action)
         (eq major-mode 'sly-mrepl-mode))
       (cons #'display-buffer-window nil)    
       nil))
```

PR #200, opened 2018-11-15
Update sly-tramp to be compatible with emacs 26

fisxoj:
This updates `sly-tramp.el` to work with emacs 26, as taken from slime!

The fix is taken from slime/slime@114bc26.


fisxoj:
It looks like these changes have already made it into sly through other means via 0e1ae127486cad4229556ff4ff3537edc9c24de5.

PR #198, opened 2018-11-14
Generalize sly tramp and add a docker-tramp contrib

fisxoj:
Hi Jão!

Thanks for your help figuring out how to run slynk in a docker container.  To continue along those lines, I was looking into ways to get the filename translations to work, too.  I haven't done a lot of elisp, nor have I contributed to sly before, so I wanted to open this as a proof of concept and ask for some helpful criticism on the elisp.  Also, would you accept this as a contrib at all?

Some questions I have:
* What will happen if the user doesn't have `docker-tramp` installed?  Will the contrib fail to load silently? (That sounds like the best case, to me)
* Do you mind changing the `car` of each entry of `sly-filename-translations` to a predicate function instead of a regex?  I could add a special case for strings if we should worry about backwards compatibility.
* The calls to `tramp-make-tramp-file-name` in `sly-tramp.el` don't seem to have enough arguments on my emacs (`GNU Emacs 26.1 (build 1, x86_64-redhat-linux-gnu, GTK+ Version 3.23.2)
 of 2018-08-13`).  Is that a bug?

Thanks!

PR #206, opened 2018-12-17
Tramp filename translator was fixed

svetlyak40wt:
When using SLY's tramp contrib, you can edit not only files opened
via tramp but also local sources. Before this patch, in the latter case,
Emacs printed "Tramp file name: /Users/art/projects/..." error.

Now, tramp contrib can work both with the tramp and plain filenames.


svetlyak40wt:
Solution, suggested in https://github.com/joaotavora/sly/pull/198 is better.

PR #223, opened 2019-04-23
Bind `*macroexpand-printer-bindings*` during print, not macroexpansion

Zulu-Inuoe:
Fix for #222


Zulu-Inuoe:
Sorry. Ignore that second commit. I didn't mean to have it be included in this PR.

PR #376, opened 2020-11-17
Backport sly-call-defun from SLIME.

Ambrevar:
* sly.el (sly-call-defun): New interactive function.
(sly-mode-map): Bind C-c C-y to sly-call-defun.

### DIFF
--- a/sly.el
+++ b/sly.el
@@ -552,6 +552,7 @@ interactive command.\".")
     (define-key map (kbd "C-c M-m") 'sly-macroexpand-all)
     ;; Misc
     (define-key map (kbd "C-c C-u") 'sly-undefine-function)
+    (define-key map (kbd "C-c C-y") 'sly-call-defun)
     map))
 
 (defvar sly-editing-mode-map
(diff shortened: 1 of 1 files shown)


Ambrevar:
Questions:

- Is (sly-current-package) the equivalent of `(slime-lisp-package)`?
- Is

```lisp
(switch-to-buffer-other-window
                  (sly-mrepl--find-buffer (sly-current-connection)))
```

the right way to switch to the REPL?

Ambrevar:
Closes #186.

joaotavora:
The functionality is interesting, and correctly defined.  But the implementation you propose is brittle because Lisp has macros. This cannot be done in Elisp alone, it needs Common Lisp support, that macroexpands enough until the underlying top-level form can be matched.

I'd propose you make this in a contrib. Maybe the contrib can start with the naive all Elisp solution, and eventually evolve into a proper solution. This kinds of things make interesting ELS talks IMO ;-)

Ambrevar:

> The functionality is interesting, and correctly defined.  But the implementation you propose is brittle because Lisp has macros. This cannot be done in Elisp alone, it needs Common Lisp support, that macroexpands enough until the underlying top-level form can be matched.

I'm confused, why doesn't it work because of macros?  All this does is
insert the current function, sayy FOO, in the REPL as

--8<---------------cut here---------------start------------->8---
CL-USER> (foo )
--8<---------------cut here---------------end--------------->8---

Am I misunderstanding something?

Besides, this is not my implementation, it's an almost exact copy-paste
of SLIME's own version ;)

P.S.: What is ELS?

joaotavora:
Consider
```
SLYNK-COMPLETION> (defmacro defhello (name &rest things)
                    `(defun ,name () (print 'hello) ,@(loop for thing in things collect `(print ',thing))))
DEFHELLO
SLYNK-COMPLETION> (defhello hi how are you doing)
HI
SLYNK-COMPLETION> (hi)

HELLO 
HOW 
ARE 
YOU 
DOING 
DOING
```
Now, if `defhello` and `hi` were defined in a source file, and you place the point inside `hi`, does `slime-call-defun` insert a call to `(hi)` in the REPL?

joaotavora:
> P.S.: What is ELS?

European Lisp Symposium

Ambrevar:
> Now, if `defhello` and `hi` were defined in a source file, and you place the point inside `hi`, does `slime-call-defun` insert a call to `(hi)` in the REPL?

Indeed this does not work.

So should I move this to an out-of-repo contrib, or can we leave this in
a (possibly non-default) contrib in this repo?

mooseyboots:
this is a really convenient function. i'm not sure why a fork of slime wouldn't include it? the slime way is to just not handle macros, by checking if top level form is one or not. it's still super useful.

joaotavora:
> Im not sure why a fork of slime wouldn't include it? 

Are the reasons I gave not clear enough?

> the slime way is to just not handle macros, by checking if top level form is one or not.

How can it know this from Elisp alone?

> it's still super useful.

Then make it a contrib!

vindarel:
> you place the point inside (defhello hi ...), does slime-call-defun insert a call to (hi) in the REPL?

it doesn't, granted, but the feature is useful for many day-to-day needs (and I use it all the time): it works in a defun, a defmacro, a defparameter/defvar/constant, in a defclass by inserting a make-instance, it inserts a fully-qualified package name if required… +1 to have this and research the custom macro case later, if needs be.

PR #381, opened 2020-11-19
Handle special characters when setting directory.

Ambrevar:
* sly.el (sly-change-directory): Wrap directory with
slynk-backend:filename-to-pathname.
* contrib/sly-mrepl.el (sly-mrepl-set-directory): Same.


Ambrevar:
Closing in favour of #379.

PR #512, opened 2022-05-13
Allow READ-FROM-MINIBUFFER-IN-EMACS to return an empty string

daewok:
The docstring for READ-FROM-MINIBUFFER-IN-EMACS says that an empty string is
returned when a user enters nothing. Prior to this commit,
SLY-READ-FROM-MINIBUFFER was called in such a way that an empty result wasn't
allowed.

### DIFF
--- a/sly.el
+++ b/sly.el
@@ -4152,7 +4152,7 @@ the display stuff that we neither need nor want."
 
 (defun sly-read-from-minibuffer-for-slynk (thread tag prompt initial-value)
   (let ((answer (condition-case nil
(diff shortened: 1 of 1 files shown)


daewok:
It kind of looks like the `t` that was already present in the call to `sly-read-from-minibuffer` may have been intended for the `allow-empty` optional argument, but instead it's consumed by the `hist` optional argument.

So I'm unsure if the correct fix is `(sly-read-from-minibuffer prompt initial-value t t)` (history is not saved) or `(sly-read-from-minibuffer prompt initial-value nil t)` (history is saved).

PR #520, opened 2022-07-12
Remove needless quote

syohex:
If this value is set via `customize`, then emacs generates the following code(double quoted symbol). This patch fixes it.

```lisp
(custom-set-variables
 '(sly-command-switch-to-existing-lisp ''always))
```


aadcg:
@syohex I pushed your commit, see b57e32049a283ac5390178e46877752db83b7d0b. Thanks.

PR #525, opened 2022-07-30
Documentation: incorrect *sly-xref* buffer commands

santiagopim:
* doc/sly.texi (cross-referencing): The `RET` and `Space` keys are
switched in the `*sly-xref*` buffer commands section of the manual.
Also the function names seems wrong, with `sly-show-xref` instead of
`sly-xref-show`. Finally, the navigation keys `n` and `p` where
missed.

Hope I've not missunderstood. Best regards and thank you for your projects.

### DIFF
--- a/doc/sly.texi
+++ b/doc/sly.texi
@@ -1181,13 +1181,13 @@ In the resulting @code{*sly-xref*} buffer, these commands are
 available:
 
 @table @kbd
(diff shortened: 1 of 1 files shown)


joaotavora:
Thanks for the changes, this is very appreciated. I'll look into this and
get back to you as soon as possible.

João

PR #553, opened 2023-01-02
Minor edits to manual.

rpgoldman:
Spelling and grammar corrections; no content changes.

### DIFF
--- a/doc/sly.texi
+++ b/doc/sly.texi
@@ -260,7 +260,7 @@ environment.  Many of the keybindings and interface concepts used to
 interact with Emacs's Elisp machine are reused in @SLY{} to interact
 with the underlying Common Lisp run-times.  Emacs makes requests to
 these processes, asking them to compile files or code snippets; deliver
-introspection information various objects; or invoke commands or
+introspection information about various objects; or invoke commands or
 debugging restarts.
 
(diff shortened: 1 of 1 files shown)


doc/sly.texi
```diff
@@ -287,7 +287,7 @@ closer to Emacs' own.
 
 To understand the differences between the two projects read
 @SLY{}'s @uref{https://github.com/joaotavora/sly/blob/master/NEWS.md,,NEWS.md}
-file.  For a hand-on approach to these differences you might want to
+file.  For a hand-on approach to these differences you might want to read
```
port19x, line 290:
```suggestion
```

This is unnecessary due to how references work in `info-mode`.
`info-mode` inserts a literal "see" before any node references, see attached screenshot.
image

port19x (changes requested):

doc/sly.texi
port19x: If other info systems handle this differently, let me know

rpgoldman: This is fixed in the current version (the word "read" no longer appears).

rpgoldman:
@joaotavora Any chance of merging this? I don't think any of it's controversial, and since it's all in `sly.texi` merging should be trivial.

History is probably not of general interest, so suggest squash-merging.

PR #558, opened 2023-01-06
Copy the returned all-completions list

### LINKED ISSUE #557: Sly's cached completion table list is being destructively modified by Corfu
With [Corfu](https://github.com/minad/corfu) enabled in common-lisp mode when completing at point, when the first candidate is equal to prompt, Corfu errors out with:
`Error in post-command-hook (corfu--post-command): (args-out-of-range 0 0)`.

This can be observed simply by trying to complete after `car` or `not`

I tracked this to the function [`corfu--move-to-front`](https://github.com/minad/corfu/blob/677933a8cd259a166d8a24d84876bbcdb18923b5/corfu.el#L527-L532) which uses `delq` on the `list` which references the cached Sly's completion table, effectively setting the first element to `nil`. See minad/corfu#285 for the reference.

As Corfu's maintainer kindly pointed out in the above mentioned issue, this seems to be the default Emacs behavior, that returned completion list can be destructively modified by a completion UI. For reference, this can be seen in the function `completion-all-sorted-completions` in `minibuffer.el`:
https://github.com/emacs-mirror/emacs/blob/15fc7b3cde92e420f48dfe188251e6af4d832af5/lisp/minibuffer.el#L1640-L1661

The easiest solution might be to return a copy of the cached completion candidate list.

Thank you for your time!

lapinskim:
Fixes #557

Copy the returned all-completions list so that it could be later destructively modified by the completion UI.

### DIFF
--- a/lib/sly-completion.el
+++ b/lib/sly-completion.el
@@ -262,12 +262,13 @@ ANNOTATION) describing each completion possibility."
                        (display-sort-function . identity)
                        (category . sly-completion)))
           ;; all completions
-          (`t (car (all)))
+          (`t (copy-sequence (car (all))))
           ;; try completion
           (`nil (try))
           (`(try-completion . ,point)
            (cons 'try-completion (cons string point)))
-          (`(all-completions . ,_point) (cons 'all-completions (car (all))))
+          (`(all-completions . ,_point) (cons 'all-completions
+                                              (copy-sequence (car (all)))))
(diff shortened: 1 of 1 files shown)


q3cpma:
Sorry if I make some annoying noise, but this seems to be a straightforward fix without much to discuss to me.

aadcg:
I'm ok with this change.

PR #563, opened 2023-01-19
Use WITH-STANDARD-IO-SYNTAX for reading literals

lukego:
Add robustness to symbol lookups of the form:

    (read-from-string "a-package::a-symbol")

by giving them stable reader settings:

    (with-standard-io-syntax ...)

so that they work the same independent of the reader settings in the image.

(I noticed that SLY compile commands stopped working when I was using named-readtables and :PRESERVE case. SLYNK would start looking for symbols with lowercase symbol-names.)


lukego:
Sorry this patch needs a respin. I can see from the diff that I missed some closing parens.

Issue #567, opened 2023-02-10
[feature request] pretty evaluation overlay


bo-tato:
Emacs lisp with [eros.el](https://github.com/xiongtx/eros) and CIDER for clojure both show evaluation results in a pretty box next to the expression, which is easier to look at then looking down at the messages and back. This would be a nice addition for sly. [This comment](https://www.reddit.com/r/emacs/comments/bi4xk1/evaluation_overlays_in_slime_for_common_lisp/) shows how to do it with slime, I changed it slightly and it works with sly:
```
(defun sly-eval-sexp-overlay ()
  (interactive)
  (eros--eval-overlay
   (let ((result (s-trim (sly-eval `(slynk:pprint-eval ,(sly-sexp-at-point))))))
     (cond
      ;; number or string, eval it
      ((-contains-p '(?0 ?1 ?2 ?3 ?4 ?5 ?6 ?7 ?8 ?9 ?\")
                    (string-to-char (substring result 0 1)))
       (eval (car (read-from-string result))))

      ;; eg #\\Newline
      ((s-starts-with-p "#\\" result) result)

      ;; cf http://www.lispworks.com/documentation/HyperSpec/Body/02_dh.htm
      ;; eg #p"arst" => "#p arst"
      ((s-starts-with-p "#" result)
       (format "%s %s"
               (substring result 0 2)
               (substring result 3 (- (length result) 1))))

      ;; maybe remove NIL from the end
      ((s-ends-with-p "\nNIL" result) result)

      ;; list
      ((and (s-starts-with-p "(" result)
            (s-ends-with-p ")" result))
       (eval (car (read-from-string (format "(quote %s)" result)))))

      ;; spaces, give up
      ((s-contains-p " " result) result)

      ;; symbol
      (t (eval (car (read-from-string (format "(quote %s)" result)))))
      ))
   (point)))
```
I don't really know any emacs lisp yet but this seems a simple enough task for me to learn, if you're interested in integrating it in sly by default I can try and make a PR request for it. Maybe the right place to add it would be to in `sly-display-eval-result` check if `eros.el` is present and if so use that rather than or in addition to the standard `message`?

bo-tato:
this doesn't print somethings properly and is also insecure if you are developing a program that prints untrusted data to standard-out (it is passing output to elisp eval). I'm now using the simpler:
```
(defun sly-eval-sexp-overlay ()
  (interactive)
[...]

PR #590, opened 2023-04-17
fix bug separating multiple values when printing

bo-tato:
`slynk-pprint-values` forgets to pass the stream to `terpri` when it prints a newline between values, so when you eval in emacs for example `(parse-integer "243")` it shows as 2433, with this fix it will print correctly:
```
243
3
```

### DIFF
--- a/slynk/slynk.lisp
+++ b/slynk/slynk.lisp
@@ -2169,7 +2169,7 @@ If STREAM is nil, use a string"
            (print-all (s)
              (loop for o in values
                    do (print-one o s)
-                      (terpri))))
+                      (terpri s))))
     (with-bindings *slynk-pprint-bindings*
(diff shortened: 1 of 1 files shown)


bo-tato:
ah true, I got it using the eval in overlay code from https://github.com/joaotavora/sly/issues/567
which is running `(sly-eval \`(slynk:pprint-eval ,(sly-sexp-at-point)))`
that overlay code has I think a few edge cases where it doesn't display things right, but unless I'm missing something this one isn't the fault of my `sly-eval-sexp-overlay`. It seems `C-x C-e` doesn't call `slynk-pprint-values`. Without the patch `slynk-pprint-values` returns the string "2433", after the patch it returns the string:
```
243
3
```
with the newline as I think is intended

PR #593, opened 2023-04-26
Always search for the full package name in existing IMPORT-FROMs

tdrhq:
.. As opposed to just looking at prefixes. Previously, If you had an existing defpackage like this:

(defpakage :foo
  (:import-from :bar/car
                :find)))

And you imported bar::position, you'll end up with something ugly like this:

(defpackage :foo
  (:import-from :bar
                :position/car
                :find))

### DIFF
--- a/contrib/sly-package-fu.el
+++ b/contrib/sly-package-fu.el
@@ -350,10 +350,15 @@ symbol in the Lisp image if possible."
 ;;
 
 (defun sly-package-fu--search-import-from (package)
+  "Moves point to the end of the package name corresponding to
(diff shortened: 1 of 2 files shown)


tdrhq:
This is a pretty common occurrence if you use single-package-per-file style, since packages have a hierarchical structure to them.

(PS. My next PR will be a bit more interesting, and you might like it. I have some patches locally where I can do C-c C-i on an symbol without the package specified, and it shows you a list of options in ido-mode to pick from. It greatly simplifies life, and I'll try sending another PR with that feature once this PR goes in.)

PR #618, opened 2023-09-13
sly-eval-print: Don't hardcode comment size

aartaka:
I'm often using `sly-eval-print` for a kind of literate programming workflow of evaluating forms one after the other and printing results inline. One particular thing that makes this workflow uncomfortable is that `sly-eval-print` prints the result with a single colon. This single colon is:
- Often improperly indented by SLY itself.
- Is against the convention that one-semicolon comments are strictly the inline ones. 
  - Block comments (two semicolons) make more sense for result printing with line break.
- And then it's hard-coded, which makes it much less customizable by user. `comment-region` has a lot of customizations, so the result of  `sly-eval-print` might be really far from what one might expect in their code comments.

This PR removes the hard-coded prefix arg, so that `sly-eval-print` comments are more customizable, consistent, and semantically correct.

### DIFF
--- a/sly.el
+++ b/sly.el
@@ -4213,7 +4213,7 @@ kill ring."
                              ""
                            " => ") value)
           (unless string-or-comment-p
-            (comment-region start (point) 1)))))))
+            (comment-region start (point))))))))
 
 (defun sly-eval-save (string)
   "Evaluate STRING in Lisp and save the result in the kill ring."


joaotavora:
This was rejected and reverted.  See reason [here](https://github.com/joaotavora/sly/commit/6a303bae74537f9955fe08e0e28d6da2b2b4ee4e#commitcomment-180233956).

> The problem is that most people (or at least "some" people, and that includes me) use C-u C-x C-e instead in lieu of your third step. And that includes the coment in-line (which is where it makes much more sense IMO). So if you want to fix your use case, you have to provide a patch that affects only sly-eval-print-last-expression (which inserts a \n before the comment).

PR #676, opened 2025-04-26
Improve the clasp slynk backend and fix a non-standard loop macro

drmeister:
I've upgraded the clasp slynk backend so that clasp can make full use of sly's enhancements.
I also found a loop macro in search-for-recording-1 that put a while clause before some for clauses and that gives our symbolics loop macro some trouble.

### DIFF
--- a/contrib/slynk-stickers.lisp
+++ b/contrib/slynk-stickers.lisp
@@ -242,7 +242,7 @@ after.")
                                       increment)
   "Return two values: a RECORDING and its position in *RECORDINGS*.
 Start searching from position FROM, an index in *RECORDINGS* which is
-successfuly increased by INCREMENT before using that to index
+successively increased by INCREMENT before using that to index
 *RECORDINGS*."
   (loop for starting-position in `(,from ,(if (plusp increment)
                                               -1
--- a/contrib/slynk-stickers.lisp
+++ b/contrib/slynk-stickers.lisp
@@ -242,7 +242,7 @@ after.")
                                       increment)
   "Return two values: a RECORDING and its position in *RECORDINGS*.
 Start searching from position FROM, an index in *RECORDINGS* which is
-successibely increased by INCREMENT before using that to index
+successfuly increased by INCREMENT before using that to index
 *RECORDINGS*."
   (loop for starting-position in `(,from ,(if (plusp increment)
                                               -1
@@ -250,14 +250,16 @@ successibely increased by INCREMENT before using that to index
         ;; this funky scheme has something to do with rollover
         ;; semantics probably
         ;;
-        for inc in `(,increment ,(if (plusp increment) 1 -1))
+        for inc in `(,increment ,(if (plusp from) 1 -1))
         for (rec idx) = (loop for cand-idx = (incf starting-position
                                                    inc)
-                              while (< -1 cand-idx (length *recordings*))
-                              for recording = (aref *recordings* cand-idx)
-                              for sid = (id-of (sticker-of recording))
-                              unless (funcall ignore-p sid)
-                                return (list recording cand-idx))
+                              if (< -1 cand-idx (length *recordings*))
+                                do (let* ((recording (aref *recordings* cand-idx))
+                                          (sid (id-of (sticker-of recording))))
+                                     (unless (funcall ignore-p sid)
+                                       (return (list recording cand-idx))))
+                              else
+                                do (return (list nil cand-idx)))
         when rec
           return (values rec idx)))
 
--- a/slynk/backend/clasp.lisp
+++ b/slynk/backend/clasp.lisp
@@ -31,13 +31,18 @@
   (when (probe-file "sys:profile.fas")
     (require :profile)
     (pushnew :profile *features*))
-  (when (probe-file "sys:serve-event")
+  (when (probe-file "sys:src;lisp;modules;serve-event;")
     (require :serve-event)
     (pushnew :serve-event *features*))
   (when (find-symbol "TEMPORARY-DIRECTORY" "EXT")
     (pushnew :temporary-directory *features*)))
 
-(declaim (optimize (debug 3)))
+;;; Compatibility tests
+
+(eval-when (:compile-toplevel :load-toplevel :execute)
+  ;; xref support (2.4)
+  (defun clasp-with-xref-p ()
+   (with-symbol 'who-calls 'ext)))
 
 ;;; Slynk-mop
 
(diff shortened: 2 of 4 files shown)


contrib/slynk-stickers.lisp
```diff
@@ -242,22 +242,24 @@ after.")
...
-successibely increased by INCREMENT before using that to index
+successfuly increased by INCREMENT before using that to index
 *RECORDINGS*."
   (loop for starting-position in `(,from ,(if (plusp increment)
                                               -1
                                               (length *recordings*)))
         ;; this funky scheme has something to do with rollover
         ;; semantics probably
         ;;
-        for inc in `(,increment ,(if (plusp increment) 1 -1))
```
joaotavora, line 253: Here's the thing: I haven't done much Common Lisp recently, so it's hard for me to review this change.

But SLY (from master) is still heavily used around some circles as far as I know.  So I'd like to know what % you are sure this is a functionally equivalent (and fully conformant) `loop` invocation.  It seems to work fine in all implementations so far except for Clasp.  How deeply have you tested this?

slynk/backend/clasp.lisp
```diff
@@ -31,13 +31,18 @@
   (when (probe-file "sys:profile.fas")
     (require :profile)
     (pushnew :profile *features*))
-  (when (probe-file "sys:serve-event")
+  (when (probe-file "sys:src;lisp;modules;serve-event;")
```
joaotavora, line 34: Thanks very much for all your work on this file.  I won't review it, but I trust it's the right thing every time :-)

slynk/slynk-loader.lisp
```diff
@@ -136,7 +136,7 @@ operating system, and hardware architecture."
 (defun sly-version-string ()
   "Return a string identifying the SLY version.
 Return nil if nothing appropriate is available."
-  (let ((this-file #.(or *compile-file-truename* *load-truename*)))
+  (let ((this-file #.(or #-clasp *compile-file-truename* *load-truename*)))
```
joaotavora, line 141: OK.  I'm reading the explanation in the commit message and I think I get it.  But that message should be even clearer.  if this rewording is correct, use something like it instead :

```
 Clasp *compile-file-truename* will be /tmp

Clasp will set *compile-file-truename* to /tmp/something.faslclasp
the containing directory of which likely won't contain 'sly.el'.  However
*load-truename* will contain the path to slynk-loader.lisp, and 'sly.el'
should be found as a result.
```

Also, as far as I remember loading `slynk-loader.lisp` is just another "legacy" way of loading SLY.  The preferred way is to load it via ASDF.  Does this work with Clasp? Did you test it?

slynk/backend/clasp.lisp
```diff
@@ -528,19 +506,6 @@
   (declare (ignore frame))
   (core:set-breakstep))
 
-(defimplementation sldb-stepper-condition-p (condition)
```
joaotavora: Fantastic.  I love red commits :-)

slynk/slynk-backend.lisp
```diff
@@ -196,7 +196,7 @@ Backends implement these functions using DEFIMPLEMENTATION."
        (if (member ',sym *interface-functions*)
            (setq *unimplemented-interfaces*
                  (remove ',sym *unimplemented-interfaces*))
-           (warn "DEFIMPLEMENTATION of undefined interface (~S)" ',sym))
+           (warn "DEFIMPLEMENTATION of undefined interface (~S)" ',name))
```
joaotavora, line 199: I think this makes sense.  Can you give a very summarized before&after of what this change does to the logs?

contrib/slynk-stickers.lisp
```diff
@@ -242,7 +242,7 @@ after.")
                                       increment)
   "Return two values: a RECORDING and its position in *RECORDINGS*.
 Start searching from position FROM, an index in *RECORDINGS* which is
-successfuly increased by INCREMENT before using that to index
+successively increased by INCREMENT before using that to index
```
joaotavora, line 245: Thanks, good catch. I think you can squash this commit in the other one where you fix the `loop` to be conforming.  If you're not comfortable enough with Git to do that, I can easily do it for you.

joaotavora (commented):
Thank you very much for these changes!  I fairly sure the code can go in as it is, but please answer my questions before we merge.

contrib/slynk-stickers.lisp
drmeister: I tested the code I wrote with sbcl, ecl, and clasp and it works fine.

The original loop code was not conformant.  To summarize, `FOR` clauses cannot appear after `WHILE` clauses.  `FOR` clauses are `variable-clauses` and `WHILE` clauses are `termination-tests`, which are `main-clauses`, which follow `variable-clauses`.

Check out:
https://www.lispworks.com/documentation/HyperSpec/Body/m_loop.htm

It says:
```
The "extended" [loop](https://www.lispworks.com/documentation/HyperSpec/Body/m_loop.htm#loop) [form](https://www.lispworks.com/documentation/HyperSpec/Body/26_glo_f.htm#form):

loop [name-clause] {variable-clause}* {main-clause}* => result*
...
variable-clause::= with-clause | initial-final | for-as-clause 
...
main-clause::= unconditional | accumulation | conditional | termination-test | initial-final 
...
termination-test::= while form | until form | repeat form | always form | never form | thereis form 
```

slynk/backend/clasp.lisp
drmeister: Ok, thanks.  Yeah,  I'll deal with issues in clasp.lisp

slynk/slynk-loader.lisp
drmeister: Ok, how would I change the commit message?  Or can you change it.   Your version is more readable.

slynk/slynk-backend.lisp
drmeister: The DEFIMPLEMENTATIONs that I removed from clasp.lisp triggered this warning and they read:

`WARNING: DEFIMPLEMENTATION of undefined interface (NIL)`

, which wasn't useful.  After this change you will see warnings like:

`WARNING: DEFIMPLEMENTATION of undefined interface (DUMMY-IMPLEMENTATION)`

I tested this on sbcl and in clasp.

contrib/slynk-stickers.lisp
drmeister: I'm not comfortable enough with git to do this - can you?

Git and I have an uneasy relationship.

drmeister:
I think I responded to all your questions.

joaotavora:
I completely forgot about this... What a shame.  Pinging myself so I don't forget.

slynk/slynk-loader.lisp
aadcg: Reworded as per @joaotavora.

contrib/slynk-stickers.lisp
aadcg: Squashed.

aadcg:
@drmeister I've pushed your work with the minor suggested changes by @joaotavora.

No regressions observed, as per `LISP=sbcl make check`. 

The test suite for `clasp` (via `LISP=clasp make check`) fails, but that is orthogonal to this PR.

Thanks.

PR #678, opened 2025-05-01
sly-sexp-at-point: error only with both `interactive` and `errorp`

Tevo45:
Seems like that was what was intended, but it ended up only testing for `interactive` twice instead.


joaotavora:
Thanks! That was definitely the intention.

PR #689, opened 2025-08-12
Fix #688: map sly keys using ctl-x-map instead of hardcoding C-x

### LINKED ISSUE #688: Eager macro-expansion failure: (error "Key sequence C-x C-a starts with non-prefix key C-x")
I have the `C-x` prefix map remapped to another key, while `C-x` I use as a simple binding to kill region or line.

Apparently sly tries to bind using something like `C-x <key>`, disregarding completely `ctl-x-map` and the users' preference. I suggest fixing that, I will open a PR with the fix soon, whenever I find the macro in which the key sequences are bound.

filipencopav:
This PR makes sly default keybindings more flexible in case a user has changed the keybindings for their `ctl-x-map` and uses `C-x` for their own usecase. Fixes #688.

### DIFF
--- a/sly.el
+++ b/sly.el
@@ -507,6 +507,12 @@ any command into it using
 Where \"k\" is the key to bind and \"sly-command\" is any
 interactive command.\".")
 
+(defun sly--keymapped-kbd (keymap keys &optional default)
+  (let ((keys-list (where-is-internal keymap)))
+    (if (null keys-list)
+        (concat (kbd default) (kbd keys))
+      (concat (car keys-list) (kbd keys)))))
+
 (defvar sly-prefix-map
   (let ((map (make-sparse-keymap)))
     (define-key map (kbd "C-r")   'sly-eval-region)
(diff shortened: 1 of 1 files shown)


joaotavora:
Does this really need to be this complicated? Why not put just `ctl-x-map` instead of the hardcoded "c-x"? @monnier is there no better way?

filipencopav:
@monnier @joaotavora sorry for the tag, any progress on this PR?

joaotavora:
Sorry, I completely misunderstood the issue.  AFAICT there is nothing to fix there. Sly doesn't bind `C-x` anything.  IT binds `C-c C-x` things. Bindings like `C-c C-x t` are completely normal in Emacs land.  I'm sorry it clashes with the Windows C-x C-c C-v things.  

If it helps I can tell you that when I started using Emacs I was obsessed myself with keeping this CUA muscle memory but then I felt liberated completely by letting it go.

PR #690, opened 2025-09-27
Fix file:// URL example in common-lisp-hyperspec-root docstring

op3r470r:
The file:// URL example in the common-lisp-hyperspec-root docstring was missing a slash.


aadcg:
Thanks @op3r470r.

PR #703, opened 2026-03-24
Change simple-break default message to "Interrupt from client"

### LINKED ISSUE #702: Make simple-break interrupt message configurable
## Summary

`simple-break` in `slynk/slynk.lisp` hardcodes the default datum as `"Interrupt from Emacs"`:

```lisp
(defslyfun simple-break (&optional (datum "Interrupt from Emacs") &rest args)
```

This message is displayed when a client sends `:emacs-interrupt` to break into a running evaluation. Clients that aren't Emacs (e.g. browser-based REPLs, CLI tools) see a confusing "Interrupt from Emacs" message in the debugger.

## Proposal

Add a `*interrupt-message*` special variable (defaulting to `"Interrupt from Emacs"` for backward compatibility) that `simple-break` uses as its default datum. Clients could then customize the message:

```lisp
(setf slynk:*interrupt-message* "Keyboard interrupt")
```

## Context

I'm building [ICL](https://github.com/atgreen/icl), a browser-based Common Lisp REPL that uses Slynk as its backend protocol. I just added Ctrl-C interrupt support, and it works great — but users see "Interrupt from Emacs" in the debugger output, which is confusing since there's no Emacs involved.

atgreen:
## Summary
- Change the hardcoded `"Interrupt from Emacs"` default datum in `simple-break` to `"Interrupt from client"`
- Non-Emacs clients using Slynk (browser-based REPLs, CLI tools) no longer see a confusing Emacs-specific message in the debugger

Closes #702

🤖 Generated with [Claude Code](https://claude.com/claude-code)


aadcg:
@joaotavora, OK to merge?

joaotavora:
No.  The commit  doesn't comply with CONTRIBUTE.md

PR #704, opened 2026-05-05
slynk: do not assume `readtable-case` to be `:upcase`

rodentrabies:
This does not affect standard Lisps and is useful for working with Allegro Common Lisp's `mlisp` which uses `:preserve`.

I've been using this patch for 5 years in all my personal development environments, but only now decided to actually submit it.


joaotavora:
I think this makes sense, but it's been so long since I've used Lisp.  What does slime do here?  And isn't it slightly nicer to use uninterned symbols `#'thing` ?  Let's try my luck and and call @luismbo here for an opinion.

rodentrabies:
Agreed on the uninterned symbols. As for SLIME's approach, the code differs a lot and I haven't used SLIME in a long time. Will try to figure out the difference in the next couple of days.

rodentrabies:
Any update on this? I think this PR is simple enough to be merged in even if there's a chance for a more general solution in the future. But I'm fine with using my fork for now for my ACL-related work.

PR #716, opened 2026-08-19
Add CL-Amiga (clamiga) SLYNK backend

mdbergmann:
[clamiga](https://github.com/mdbergmann/cl-amiga) is a from-scratch Common Lisp
  implementation (a C bytecode interpreter with an m68k JIT) targeting AmigaOS 3+, MorphOS
  and POSIX hosts. This adds SLY support for it.

  * `slynk/backend/clamiga.lisp` (new) — TCP server, threads and message passing,
    locks, compilation, arglists/describe, definition finding, SLDB.
  * `slynk/slynk.asd` — `#+clamiga (:file "clamiga")` in the backend module, so
    the default ASDF init works with no custom `:init`.
  * `slynk/slynk-loader.lisp` — register the backend, the `:clamiga` feature and a
    version string, so the slynk-loader path works too. Also fixes a latent paren
    bug in the `#+cormanlisp` override, where `FORCE-OUTPUT` was a bogus argument
    to `LOAD`.

  `MAKE-LOCK` returns a *recursive* lock: SLY's Gray output stream re-enters its
  own buffer-write lock (`STREAM-WRITE-STRING` → `STREAM-FINISH-OUTPUT`), and
  clamiga's non-recursive mutex self-deadlocks there. CCL's locks are recursive
  and SLY assumes that.

  Verified end-to-end: both the ASDF init and the slynk-loader path start a server
  and round-trip `slynk:connection-info`. No other implementation is affected,
  apart from the `#+cormanlisp` paren fix.

### DIFF
--- a/slynk/backend/clamiga.lisp
+++ b/slynk/backend/clamiga.lisp
@@ -1,0 +1,508 @@
+;;;; -*- indent-tabs-mode: nil -*-
+;;;
+;;; slynk-clamiga.lisp --- SLY backend for clamiga (CL-Amiga).
+;;;
+;;; This code has been placed in the Public Domain.  All warranties
+;;; are disclaimed.
+;;;
+;;; clamiga is a from-scratch Common Lisp implementation (a C bytecode
+;;; interpreter with an m68k JIT) targeting AmigaOS and POSIX hosts.  Its
+;;; SLY-facing facilities live in three packages:
+;;;
+;;;   MP   threads, locks, condition variables   (src/core/builtins_thread.c)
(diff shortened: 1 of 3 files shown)


joaotavora:
Sounds cool (although name sounds a bit dangerously close to "clamidia" :-) )

Do you foresee lots of changes to `slynk/backend/clamiga.lisp`  in the near future?  Also how much of it was LLM-written?  (fine if 100%, I just want to know)

mdbergmann:
Clamidia, oh, I hope people don't mistake it. :)

It should be quite stable for this part, so I don't expect many changes.
It was completely vibe coded.

PR #719, opened 2026-09-18
slynk: Allow listening on a unix domain socket

fade:
Two commits: a descriptor leak in the SBCL backend's `create-socket`, and
then a new option to have Slynk listen on a unix domain socket.

### 1. `slynk/sbcl: Close the listen socket if setting it up fails.`

`create-socket` makes a socket and then configures, binds and listens on
it. Any of those three can signal (a bind on a port already in use being
the ordinary case) and the socket was left unreferenced with its
descriptor still open.

`socket-quest` offers a restart for exactly that error so the user can try
another port, so it is not a one-off: it costs one descriptor per attempt.
Twenty refused binds against the unmodified code leak twenty descriptors,
counted from `/proc/self/fd` before and after.

It is unrelated to the feature below, so it is a separate commit.

### 2. `slynk: Allow listening on a unix domain socket.`

Pass a pathname as `create-server`'s `interface` argument:

```lisp
(slynk:create-server :interface #p"/run/slynk/repl.sock")
```

I want the listener reachable by filesystem path and gated by the
permissions on that path, rather than by whatever can reach the loopback
address. It forwards over ssh without opening a loopback port on the
remote side: `ssh -L 4005:/run/slynk/repl.sock host`, then `M-x
sly-connect` to the local port. Nothing changes on the Emacs side.

A unix socket has no port number, so `local-port` returns nil for one and
`setup-server` designates the server by its path instead. That designator
is what `create-server` returns and what `stop-server` takes, so the
sentinel's lookup compares with `equal` rather than `eql`. Closing such a
listener unlinks its file so the path can be bound again.

**Permissions.** `*unix-socket-mode*` defaults to `#o600` and
`create-server` takes a `mode` argument to override it, or nil to leave the
mode to the process umask. It is applied between the bind and the listen,
so nothing can connect to the socket while its mode is still the umask's.
The path is visible with the umask's mode for the span of those two calls,
and the docstring says so rather than claiming more. I chose that over wrapping the
bind in a umask, which is process-global and would mis-permission files
written by other threads, and over binding in a private directory and
renaming into place, which would silently replace a live server's socket
where a refused bind protects it.

`mode` reaches the backend through a special that `create-server` binds,
the way it already binds `*loopback-interface*` to communicate the address.
Adding it to the portable `create-socket` lambda list would not work:
`definterface` expands to a dispatcher that passes every declared keyword
whether or not it was supplied, so the other backends would receive a
`:mode` they do not accept.

**Only the SBCL backend implements this**, and only where a unix socket
exists: the pathname branch is `#-win32`, so on Windows a pathname falls
through to `resolve-hostname` as before. The other backends are unchanged
and do not accept a pathname; the commit message says so. I have not tried
to write the parallel change for implementations I cannot test.

### Checks

Cold SBCL with the fasl cache cleared. A TCP `create-server` and a unix one
each complete a real `eval-for-emacs` round trip. I counted descriptors
before and after twenty failed attempts on each failure path: a bind
refused against a live server, and a listen that fails after a successful
bind. Both hold at zero. In the refused-bind case the incumbent's socket
file has the same inode afterwards and still answers, so a failed bind does
not disturb the server that owns the path. I read the socket modes back with
`stat` rather than assuming them: 600 by default, 660 and 700 when asked,
and 755 with `mode` nil, which is what the umask alone produced here.

I drove one of those round trips through a real forward rather than only
describing one: `ssh -N -L 127.0.0.1:47513:/tmp/slynk-ssh-forward.sock
localhost` in a separate process, then a client on the local port, which
returned `(:return (:ok 3) 1)` like the direct ones. Once I killed the
forward the local port refused the connection, while the socket itself
still answered when reached directly, so the forward is what carried it.

One thing I noticed and did not chase: `stop-server` called immediately
after `create-server` can lose a race with the `:add-server` message the
spawned serve-loop sends, and warns `No server for :PORT:`. It reproduces
on unmodified code over TCP, so it is not from these changes.


joaotavora:
Which model did you use?

fade:
I use claude in gptel to file off the really serious grammar/spelling foibles, as well as organizing commits,  messages, and PRs. Lately, some code review also.

joaotavora:
Sorry, I don't have time or resources to review this wall of text/code as it is. Can you check if this is on SLIME and, of not, why not? What do you think of this implemention?

